PR #97, opened 2016-02-19
Increases labelComponent api consistency between VictoryScatter and VictoryBar

cpb:
# Work In Progress

# Resolves
https://github.com/FormidableLabs/victory-chart/issues/96

# Todo
- [x] Make the label style merging the same between ```VictoryBar``` and ```VictoryScatter```
- [x] Update the documentation for each property with text from #95

### DIFF
--- a/src/components/victory-bar/victory-bar.jsx
+++ b/src/components/victory-bar/victory-bar.jsx
@@ -142,11 +142,19 @@ export default class VictoryBar extends React.Component {
      */
     labels: PropTypes.array,
     /**
+     * The labelComponent prop takes in an entire, HTML-complete label component
+     * which will be used to create labels for scatter to use
+     */
+    labelComponent: PropTypes.element,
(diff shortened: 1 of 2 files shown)


cpb:
I was wrong about the inconsistency with the styles. Sorry I didn't read it more carefully: https://github.com/FormidableLabs/victory-chart/blob/0c7b1855c5da4847f50b709c07b3bbf74cbf6809/src/components/victory-scatter/point.jsx#L72
vs
https://github.com/FormidableLabs/victory-chart/blob/0c7b1855c5da4847f50b709c07b3bbf74cbf6809/src/components/victory-bar/bar-label.jsx#L43

The actual difference is that ```VictoryBar``` also evaluates datum styles. Sorry for the confusion!

cpb:
see https://github.com/FormidableLabs/victory-chart/pull/98

PR #106, opened 2016-02-23
Prevent axis label collisions and overflows

tptee:
This PR uses a naive constraint system to prevent label collisions and overflows. If labels are too wide, they are truncated and given ellipses. If multiline labels are too tall, their last line is deleted until they are short enough. The constraints fail gracefully if given a font size that makes collisions/overflows inevitable.

I'm not handling line height yet, but it should be a quick fix. @boygirl do we handle line height for multiline text yet?

This PR is a little messier than I'd like it to be, so I'd love all the review I can get! cc @boygirl @coopy @alexlande @zachhale @baer @exogen @divmain

### DIFF
--- a/src/components/victory-axis/constrain-tick-labels.js
+++ b/src/components/victory-axis/constrain-tick-labels.js
@@ -20,17 +20,18 @@ const getFontWeight = (fontWeight) => {
 };
 
 const measureTicks = (props, layoutProps, tickProps) => {
-  const { fontSize, fontWeight } = props.style.tickLabels;
+  const FALLBACK_LINE_HEIGHT = 1.2;
+
+  const { fontSize, fontWeight, lineHeight } = props.style.tickLabels;
   const { fontFamily } = layoutProps.style.tickLabels;
   const tickFormat = Helpers.getTickFormat(props, tickProps);
-
   return props.tickValues.map((tick, index) => {
     const text = tickFormat(tick, index);
     const measurement = measureText({
       text: textToArray(text),
       fontFamily,
       fontSize: `${fontSize}px`,
-      lineHeight: 1.2, // TODO: where do you find this?
+      lineHeight: lineHeight || FALLBACK_LINE_HEIGHT,
       fontWeight: getFontWeight(fontWeight),
       canvas: props.canvas
     });
--- a/src/components/victory-axis/constrain-tick-labels.js
+++ b/src/components/victory-axis/constrain-tick-labels.js
@@ -1,0 +1,189 @@
+import measureText from "measure-text";
+import Helpers from "./helper-methods.js";
+
+const parsePx = (string) => parseFloat(
+  string.replace("px", "")
+);
+
+const textToArray = (text) => {
+  return typeof text === "string"
+    ? text.split("\n")
+    : text.toString().split("\n");
+};
+
+const getFontWeight = (fontWeight) => {
+  if (fontWeight) {
+    return typeof fontWeight === "function"
+      ? fontWeight() : fontWeight;
+  }
+  return "normal";
+};
+
+const measureTicks = (props, layoutProps, tickProps) => {
+  const { fontSize, fontWeight } = props.style.tickLabels;
+  const { fontFamily } = layoutProps.style.tickLabels;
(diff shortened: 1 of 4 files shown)


src/components/victory-axis/constrain-tick-labels.js
```diff
@@ -0,0 +1,189 @@
...
+  const { fontFamily } = layoutProps.style.tickLabels;
+  const tickFormat = Helpers.getTickFormat(props, tickProps);
+
+  return props.tickValues.map((tick, index) => {
+    const text = tickFormat(tick, index);
+    const measurement = measureText({
+      text: textToArray(text),
+      fontFamily,
+      fontSize: `${fontSize}px`,
+      lineHeight: 1.2, // TODO: where do you find this?
```
boygirl: I believe I grabbed this from looking at d3's axis implementation.

dandelany:
:clap: excited to check out `text-measure`, thanks @tptee

tptee:
@dandelany you got it! Would love your feedback, it's very alpha right now :smile:

boygirl:
@tptee something funny is going on with animating tick labels in this branch:
axis-animation

tptee:
Sure is! I'll investigate.

src/components/victory-axis/constrain-tick-labels.js
```diff
@@ -0,0 +1,190 @@
...
+const textToArray = (text) => {
+  return typeof text === "string"
+    ? text.split("\n")
+    : text.toString().split("\n");
+};
+
+const getFontWeight = (fontWeight) => {
+  if (fontWeight) {
+    return typeof fontWeight === "function"
+      ? fontWeight() : fontWeight;
```
coopy: Consistency: Put the `: fontWeight` statement on new line, like on line 11

test/client/spec/helpers/constrain-tick-labels.spec.js
```diff
@@ -0,0 +1,433 @@
...
+  for (const measurement of measurements) {
+    const tickValues = yield;
+    yield tickValues.map(mapper(measurement));
+  }
+};
+
+const mockMeasureTicks = (generator) => {
+  return (props) => {
+    const result = generator.next(props.tickValues).value;
+    generator.next();
```
coopy: Hmm, explain why we're assigning a `const` from the first call to `next().value`, and then calling `next()` without caring about the result?

test/client/spec/helpers/constrain-tick-labels.spec.js
```diff
@@ -0,0 +1,433 @@
...
+      const constants = Array(4).fill({
+        width: `${normalWidth}px`, height: `${height}px`
+      });
+      const measureTextGenerator = mockMeasureText([
+        [...constants, {width: `${normalWidth + 20}px`, height: `${height}px`}],
+        [...constants, {width: `${normalWidth + 10}px`, height: `${height}px`}],
+        [...constants, {width: `${normalWidth}px`, height: `${height}px`}],
+        [...constants, {width: `${normalWidth}px`, height: `${height}px`}]
+      ]);
+      measureTextGenerator.next();
```
coopy: Why are you calling `next()` here, throwing away the first result...

test/client/spec/helpers/constrain-tick-labels.spec.js
```diff
@@ -0,0 +1,433 @@
...
+      const ticks = constrainTickLabels({
+        props: defaultProps,
+        layoutProps: defaultLayoutProps,
+        tickProps: defaultTickProps,
+        measureFunc: mockMeasureTicks(measureTextGenerator)
```
coopy: ...and then using the rest of the generated results here? I don't understand generators very well yet.

tptee: Generators that take parameters in next() are super weird and require these sorts of gymnastics. Section 4.1 (here)[http://www.2ality.com/2015/03/es6-generators.html] explains some of the strangeness involved!

tptee:
We'll have a better solution for this with #135. Closing in favor of it.

PR #135, opened 2016-03-02
[WIP] Autolayout for axis

tptee:
**NOT READY FOR MERGE, EVERYTHING'S STILL GROSS**
This is the beginning of work on using a subset of Autolayout.js for axis layout. If this layout system and PR is approved, we'll move on to using Autolayout for victory-bar.

A constraint-based layout system will help us in some degree with the following issues:
#33 #106 #107 #121 #122 
FormidableLabs/victory#41

TODO
- [ ] Reimplement custom label component support for axis and tick labels
- [x] Layout axis labels
- [ ] Fix layouts for all orientations
- [ ] Fix layout for cross axis
- [ ] Avoid Autolayout.js huge bundle size by not including VFL parser code and using the kiwi solver instead of cassowary.js
- [ ] Extract AutoLayout component into a standalone module 
- [ ] Pass tests and lint
- [ ] Refactor vigorously after everything works

### DIFF
--- a/demo/components/victory-axis-demo.jsx
+++ b/demo/components/victory-axis-demo.jsx
@@ -26,12 +26,12 @@ export default class App extends React.Component {
 
   componentDidMount() {
     /* eslint-disable react/no-did-mount-set-state */
-    this.setStateInterval = window.setInterval(() => {
-      this.setState({
-        tickValues: this.getTickValues(),
-        domain: this.getDomain()
-      });
-    }, 2000);
+    //this.setStateInterval = window.setInterval(() => {
+      //this.setState({
+        //tickValues: this.getTickValues(),
+        //domain: this.getDomain()
+      //});
+    //}, 2000);
   }
 
   componentWillUnmount() {
--- a/package.json
+++ b/package.json
@@ -21,8 +21,9 @@
     "version": "builder run npm:version && git add dist && git commit -m \"Commit 'dist/' for publishing\""
   },
   "dependencies": {
-    "builder-victory-component": "~0.2.1",
+    "autolayout": "^0.5.3",
     "builder": "~2.4.0",
+    "builder-victory-component": "~0.2.1",
     "d3-scale": "^0.2.0",
     "d3-shape": "^0.2.0",
     "lodash": "^3.10.1",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "d3-scale": "^0.2.0",
     "d3-shape": "^0.2.0",
     "lodash": "^3.10.1",
+    "measure-text": "0.0.4",
     "memoizerific": "^1.5.2",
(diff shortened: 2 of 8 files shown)


tptee:
## The `<AutoLayout>` component
```es6
<AutoLayout
        width={450}
        height={300}
        container="g"
        constraints={[
          constrain("line", "left").equals(null, "left")
            .plus(layoutProps.padding.left),
          constrain("line", "right").equals(null, "right")
            .minus(layoutProps.padding.right),
          constrain("line", "bottom").equals(null, "bottom").minus(50),
          constrain("line", "centerX").equals(null, "centerX")
        ]}
      >
        {children}
      </AutoLayout>
```
The `container` prop determines what type of node to use as the wrapper for all child components (in our case with SVG, `<g />`).

The `constraints` prop allows definition of initial constraints. Child components can also define constraints in their `constraint` props.

## Children of `<AutoLayout>`

Any children of the AutoLayout component need the `viewName` prop. This is a unique identifier like `key` that elements can refer to in their constraints. Children can also have a `constraints` property that AutoLayout resolves.

`Constraints` is a convenience prop: since Autolayout doesn't depend on hierarchical relationships, the constraint tree is technically a flat array. The prop lets you organize your constraints in a hierarchical-feeling manner.

## `constrain()`
This is a "constraint builder" with a fluent interface. The native constraint representation in Autolayout.js isn't particularly expressive, so this function provides sugar over it.

Examples:
```es6
// Leave the right viewName as null to refer to the AutoLayout wrapper component
constrain("line", "left").equals(null, "left")
  .plus(layoutProps.padding.left),
constrain("line", "right").equals(null, "right")
  .minus(layoutProps.padding.right),
constrain("line", "bottom").equals(null, "bottom").minus(50),
constrain("line", "centerX").equals(null, "centerX")

constrain(tickName, "bottom").lessThanOrEqualTo(null, "bottom"),
constrain(tickName, "height").constant(5)
```

Check out https://github.com/IjzerenHein/autolayout.js/blob/master/docs/AutoLayout.md for supported attributes.

exogen:
As someone who has built lots of fluent chainable interfaces, make sure in the final implementation you return a new `Constraint` instance from every method instead of modifying and returning `this`. Chainable interfaces should be composable like so:

```
const base = constraint("line", "right");
const someConstraintA = base.equals(null, "right").minus(10);
const someOtherConstraintB = base.constant(5);
```

^ you don't want `someOtherConstraintB` to secretly mutate `someConstraintA`.

tptee:
Nice catch! Yeah, I thought that `this` stuff felt nasty...

tptee:
Currently breaking out the AutoLayout component and constraint builder into `radium-autolayout`, pending fixes to `builder-radium-component`!

briandastous:
So, this feature didn't work out?

PR #177, opened 2016-04-06
Fix victory bar transitions

divmain:
Along with the changes [here](https://github.com/FormidableLabs/victory-core/pull/44), this fixes transitions for `VictoryBar`, and corrects an oversight in `VictoryChart` that didn't cause issues with `VictoryScatter` transitions.

- Use `oldProps` for `getNewChildren` (which effected domain calculations) when rendering the exit transition.
- Gives `VictoryBar` some default transitions.
- Adds a demo of `VictoryBar` plus transitions to the demo page.

I noticed that [this line](https://github.com/FormidableLabs/victory-chart/blob/347e1ab748dc7b8c0becab72072402257ec98627/demo/components/victory-chart-demo.jsx#L187) is throwing tons of errors in the demo page - something along the lines of `getFullYear is not a function`.  I haven't checked yet whether this is happening master - it seems unrelated to the changes I've made here, but want to confirm before this is merged in.

- [ ] The accompanying `victory-core` PR should be merged and the reference updated here before this PR is merged.

bar transitions

@boygirl @kenwheeler

### DIFF
--- a/demo/components/victory-chart-demo.jsx
+++ b/demo/components/victory-chart-demo.jsx
@@ -19,7 +19,8 @@ class App extends React.Component {
       lineData: this.getData(),
       numericBarData: this.getNumericBarData(),
       barData: this.getBarData(),
-      lineStyle: this.getStyles()
+      lineStyle: this.getStyles(),
+      barTransitionData: this.getBarTransitionData()
     };
   }
 
@@ -70,6 +71,13 @@ class App extends React.Component {
     });
   }
 
+  getBarTransitionData() {
+    const bars = _.random(6, 10);
+    return _.map(_.range(bars), (bar) => {
+      return { x: bar, y: _.random(2, 10) };
+    });
+  }
+
   getScatterData() {
     const colors =
       ["violet", "cornflowerblue", "gold", "orange", "turquoise", "tomato", "greenyellow"];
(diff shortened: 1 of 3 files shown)


boygirl:
cc/ @divmain 

With these changes, and the accompanying changes in victory-core I'm still seeing some odd animation behavior in the bar chart animation.
bar-transition-bug

Also, the scatter animation is no longer animating at all...

divmain:
@boygirl are you sure you're using the correct `victory-core` branch?  Here's what the two together look like on my machine:
both-transitions

boygirl:
cc/ @divmain I checked again, nuking `node_modules` and re `npm link` ing. No good. I noticed that you have two branches `fix-transitions-for-bar-graph` and `fix-victory-bar-transitions`. Maybe the wrong one was PR'd?

boygirl:
`getFullYear is not a function` bug is not seen on master. It seems related to this change https://github.com/FormidableLabs/victory-core/pull/44/files#diff-59a48fa1294cfe45f7f6d5309ca66bd6L253 but I don't know why. As far as I can tell, the only side effect is assigning a prop `data: undefined` to victory-axis.

boygirl:
fixed by #176

PR #258, opened 2016-06-22
Fix y axis label

ebrillhart:
fixes https://github.com/FormidableLabs/victory/issues/285 

redoing some of the math that calculates x and y positioning for axis labels

PR #263, opened 2016-06-24
fix axis labels

ebrillhart:
if this works I will close the other pull request! this would fix #258 

@boygirl


ebrillhart:
@boygirl nevermind... the one with the offset would still need padding to look right 👎

ebrillhart:
@boygirl passing this back to you

PR #290, opened 2016-07-07
Add ecologyPlaygroundLoading func to docs

kylecesmat:
Will load placeholder boxes on initial render and fix https://github.com/FormidableLabs/formidable-landers/issues/23

### DIFF
--- a/docs/victory-area/docs.js
+++ b/docs/victory-area/docs.js
@@ -5,7 +5,7 @@ import Radium, { Style } from "radium";
 import * as docgen from "react-docgen";
 import { merge } from "lodash";
 import { VictoryArea, VictoryStack, VictoryGroup, VictoryScatter } from "../../src/index";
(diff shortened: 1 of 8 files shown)


kylecesmat:
cc @paulathevalley

docs/victory-area/docs.js
```diff
@@ -18,7 +18,7 @@ class Docs extends React.Component {
           playgroundtheme="elegant"
           customRenderers={appendLinkIcon}
         />
-        <Style rules={VictoryTheme}/>
+        <Style rules={merge(ecologyPlaygroundLoading, VictoryTheme)}/>
```
paulathevalley: Should this be on line 19?

paulathevalley:
1 comment to address, otherwise LGTM!

docs/victory-area/docs.js
kylecesmat: YES, good catch this is very wrong

PR #291, opened 2016-07-07
Add material theme to demos

paulathevalley:
Base branch: https://github.com/FormidableLabs/victory-chart/pull/286

Use material theme in demos instead of grayscale theme.

/cc @ebrillhart @kylecesmat @boygirl


kylecesmat:
I think material > grayscale is a good move for defaults 🚀

ebrillhart:
this theme looks so good

PR #298, opened 2016-07-14
[NO REVIEW] Add dropdown to change data sets 

paulathevalley:
Resolves https://github.com/FormidableLabs/victory-docs/issues/29 &  https://github.com/FormidableLabs/ecology/issues/33 

- Add a `<DataSource>` wrapper
- Since not all examples need a dropdown to change data sets, add an example to VictoryChart to talk about domain and scaling, where changing a data set does make sense. 

TODO:
- Add copy for domain and scale 
- Explore other potential use cases for `<DataSource>`
- Potentially refactor `<DataSource>` such that it does not require `playground_norender` ...this will require adding a child component that sets `contextTypes`, for example:
```
DataSourceChild.contextTypes = {
  dataSet: React.PropTypes.array
};
```

### DIFF
--- a/docs/victory-chart/data-source.js
+++ b/docs/victory-chart/data-source.js
@@ -1,0 +1,59 @@
+import React from "react";
+
+const dataSets = [
+  [{x: 0, y: 0}, {x: 10, y: 20}, {x: 2, y: 1}],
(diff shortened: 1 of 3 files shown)


coopy:
Thanks for posting this! Needed to peek a the code we wrote.

PR #300, opened 2016-07-18
Add Data Source Dropdowns

kylecesmat:
Some victory components would be better visualized with multiple datasets. This will add a select dropdown to the component playgrounds and let the user toggle between dataset options.

How it works:
- DataSource component takes `dataset` as a prop, and exposes the data through context to its children
- Playground component includes DataSource componentry and reference to desired dataset

Extends original work here - https://github.com/FormidableLabs/victory-chart/pull/298

This PR Changes:
- Add dataset-dropdown component to /docs
- Use the dropdown to show domain/scale changes https://github.com/FormidableLabs/victory-chart/pull/300/files#diff-50bc30efbbd4159a3c53d3dc58ea7b52R33
- Use latest Formidable-landers


kylecesmat:
Will include the dataset-dropdown from formidable-landers once it is merged & tagged

kylecesmat:
cc @paulathevalley Ready for review! CI Will fail until a new version of formidable-landers is published following https://github.com/FormidableLabs/formidable-landers/pull/115

paulathevalley:
:+1: eeeeexcellent once formidable-landers is published and CI passes!

PR #304, opened 2016-07-20
Fix/error bar animate

chilijung:
#302


boygirl:
@chilijung I'm still seeing 
error-bar-animate

chilijung:
@boygirl I think that example doesn't send a animate prop. https://github.com/FormidableLabs/victory-chart/blob/master/demo/components/victory-errorbar-demo.js#L108-L116

boygirl:
@chilijung my mistake! merging

PR #311, opened 2016-07-22
fix line, area chart transition problem

chilijung:
@boygirl I've finished the main part of improving line and area chart transition.

Before: 

before

After: 

line_animation

The only thing left is `clipPath` haven't added. I have some questions before doing further implementation.

## 1.   

Do we need `onExit` after function, because this is strange. In my opinion, I think onExit after means the data point is deleted in the chart, so it isn't exist?

## 2.   

The `onExit` and `onEnter` work different in bar type charts and line type charts, the modification I do in line type charts https://github.com/FormidableLabs/victory-core/commit/91591c6f8edf97ceadd56522f76da3fa108b7bde , we should think of a way to work prefect.

Differences: 

### Bar type charts
- When points need to exit, the points will exit first and then transition to the next state.
- When points need to enter, the transition will goes to the next state first and then the points will enter afterward.

### Line type charts
- When points need to exit, the points should transition to the next state and exit.
- When points need to enter, the points should enter first and transition.

### DIFF
--- a/demo/components/victory-errorbar-demo.js
+++ b/demo/components/victory-errorbar-demo.js
@@ -43,7 +43,7 @@ export default class App extends React.Component {
       this.setState({
         data: getData()
       });
-    }, 10000);
+    }, 3000);
   }
 
   componentWillUnmount() {
--- a/demo/components/victory-line-demo.js
+++ b/demo/components/victory-line-demo.js
@@ -115,17 +115,17 @@ export default class App extends React.Component {
       <div className="demo">
         <h1>VictoryLine</h1>
 
-          <VictoryLine
-            style={{parent: parentStyle, data: this.state.style}}
-            data={this.state.transitionData}
-            animate={{duration: 800}}
-            containerComponent={
-              <VictoryContainer
-                title="Line Chart"
-                desc="This is a line chart for displaying data."
-              />
-            }
-          />
+        <VictoryLine
+          style={{parent: parentStyle, data: this.state.style}}
+          data={this.state.transitionData}
+          animate={{duration: 800}}
+          containerComponent={
+            <VictoryContainer
+              title="Line Chart"
+              desc="This is a line chart for displaying data."
+            />
+          }
+        />
 
         <VictoryLine
           style={{parent: parentStyle, data: this.state.style}}
@@ -235,13 +235,17 @@ export default class App extends React.Component {
           theme={VictoryTheme.material}
         >
           <VictoryLine
-            style={{parent: parentStyle}}
-            data={this.state.data}
-            label="Hello"
-            x={"x"}
-            y={(d) => (d.y + 15)}
+            style={{parent: parentStyle, data: this.state.style}}
+            data={this.state.transitionData}
+            animate={{duration: 1500}}
+            containerComponent={
+              <VictoryContainer
+                title="Line Chart"
+                desc="This is a line chart for displaying data."
+              />
+            }
           />
-      </VictoryChart>
+        </VictoryChart>
       </div>
     );
   }
--- a/demo/components/victory-line-demo.js
+++ b/demo/components/victory-line-demo.js
@@ -102,7 +102,7 @@ export default class App extends React.Component {
         transitionData: this.getTransitionData(),
         style: this.getStyles()
       });
-    }, 2000);
+    }, 3000);
   }
 
   componentWillUnmount() {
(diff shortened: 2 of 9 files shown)


boygirl:
@chilijung this change also has weird consequences for scatter charts

chilijung:
This fixed is done!!!! I finished everything that we discuss yesterday! 🎉 

area_animate
line_animate

## Accomplished

- Added `ClipPath` component in component `helpers` folder, add a `clipPathComponent` prop in VictoryLine and VictoryArea, so we can change to adapt into React Native.
- Add `VictoryTransition` and `VictoryContinuousTransition` component for continuous and discrete chart transitions, also added warning when using the wrong transition in chart component.

Just some tests to test the implement haven't added.

boygirl:
@chilijung One problem I notice... the axis no longer animates for continuous transitions. Compare with how it works for discrete data
correct-axis-animation

boygirl:
Other weird behavior with the clipPath. Maybe this is because we have responsive container (viewBox), but strangely I think it's only happening on charts that just animate without any entering / exiting nodes, but I'm not certain. I would need to test more. 

weird-clip-path

chilijung:
hmm, @boygirl I think the axes animation and clipPath cutting container could be fixed. I'll look into it later today.

boygirl:
I wonder... if you have continuous and discrete data like a line and bars on the same chart and both are animating / transitioning what will happen with the domain? Maybe this is the same problem that is happening with the axes because they will act like discrete data.

chilijung:
the only thing different between the two is https://github.com/FormidableLabs/victory-core/pull/87/files/91591c6f8edf97ceadd56522f76da3fa108b7bde#diff-70416718ed309d68a94f53c5a9bc1b3cL90 , in continuous chart, it should remain the previous domain when the state `nodesWillExit` is true, and update the domain later

chilijung:
@boygirl the two problem you mentioned earlier is fixed!

here is a demo, that how it would animate when both lines and bars are exist.

animate

boygirl:
@chilijung So close, but now the entering bars start outside of the chart area.

boygirl:
@chilijung the entrance transitions work perfectly, but during the exit transitions, the axis gets out of sync with the other children
exittransitions

boygirl:
Actually... looking at this example with mixed children, it seems like we might just have to make a choice about what domain to use, since the axis animation seems to line up with the continuous animations.  I notice that the bars start to show up outside the domain thought too... Maybe we need clipPath there too?
bar-entrance

boygirl:
I have a thought... what if the transition for line / area is actually the same exactly as the other types, and the clipPath starts at the old domain and moves to the new domain. So instead of the line always going to the edge of the chart, it would start sometimes in the middle of the chart and move to the edge. Then by default the clipPath can always cut off stuff outside the domain so that when elements render and slide in they are never outside the axes. Does that make sense?

chilijung:
> @chilijung the entrance transitions work perfectly, but during the exit transitions, the axis gets out of sync with the other children

@boygirl for the axes, they are not sync with anyone. They are just doing their own animations, I've remove the parentState so the lines, bars, axes will animate independently.

As we are mounting the newChildren in the same time.
https://github.com/FormidableLabs/victory-chart/blob/fix/transition/src/components/victory-chart/victory-chart.js#L366-L376
and pass animate props the same to children.
https://github.com/FormidableLabs/victory-chart/blob/fix/transition/src/components/victory-chart/victory-chart.js#L326

Do you think this is ok? If I have them animate independently? I think it is hard axes to consider sync with continuous and discrete chart the same time, since we are doing continuous and discrete animation in the opposite way.

>  I notice that the bars start to show up outside the domain thought too... Maybe we need clipPath there too?

Yes, we could!

> I have a thought... what if the transition for line / area is actually the same exactly as the other types, and the clipPath starts at the old domain and moves to the new domain. So instead of the line always going to the edge of the chart, it would start sometimes in the middle of the chart and move to the edge. Then by default the clipPath can always cut off stuff outside the domain so that when elements render and slide in they are never outside the axes. Does that make sense?

Yeah, that would work but I think it would look like the old way?

[!old way](https://camo.githubusercontent.com/7d84e9fe480f29909337e4891201b4f5630474a0/687474703a2f2f692e696d6775722e636f6d2f694134656963622e676966)

If we want to make it look smooth we should keep the new way.

PR #314, opened 2016-07-22
Docs: Move Style component and spellcheck for the props table

paulathevalley:
Resolves https://github.com/FormidableLabs/victory-docs/issues/38

- Move repeated `<Style rules={VictoryTheme />` to `app.js` and delete from every `docs.js` 
- Spellcheck: `@example` should be `@examples` 
- `VictoryTheme.grayscale` no longer exists, so replaced it with `VictoryTheme.material` and removed the link (which looked kinda weird in the Props Table anyway, and with a Theme page it should be easier to see it in full glory)  

/cc @ebrillhart @kylecesmat @bmathews

### DIFF
--- a/docs/app.js
+++ b/docs/app.js
@@ -1,6 +1,8 @@
 import React from "react";
 import { Link } from "react-router";
-import { StyleRoot } from "radium";
+import { StyleRoot, Style } from "radium";
+import { VictoryTheme } from "formidable-landers";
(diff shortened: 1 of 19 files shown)


bmathews:
:+1:

kylecesmat:
Looks great! one style tag to rule them all!

PR #322, opened 2016-07-28
Minor Example Changes

kylecesmat:
This should fix a few examples that looked broken in the preview

cc @paulathevalley


paulathevalley:
:guitar: :rocket:

PR #333, opened 2016-08-04
Fix/transition with curtain method

chilijung:
@boygirl ! fix transition with curtain method 

DEMO!!!!

animate
line

This work perfect with bars and axes! If it also looks good to you. I'll implement in area chart!


chilijung:
related https://github.com/FormidableLabs/victory-core/pull/92

boygirl:
@chilijung for me with the victory-core linked the demo is very slow with lots of console errors. It's hard to tell if it's working right

src/components/victory-area/victory-area.js
```diff
@@ -34,12 +37,39 @@ export default class VictoryArea extends React.Component {
   static defaultTransitions = {
     onExit: {
       duration: 500,
-      before: () => ({ y: 0, yOffset: 0 })
+      before: () => ({ y: 0, yOffset: 0 }),
+      beforeClipPathWidth: (data, child, exitingNodes) => {
+        const filterExit = filter(data, (datum) => { return !exitingNodes[datum.x]; });
+        const xVals = filterExit.map((datum) => {
+          return child.type.getBaseProps(child.props).all.data.scale.x(datum.x);
```
boygirl: don't use `getBaseProps`

src/components/victory-line/victory-line.js
```diff
@@ -40,12 +43,39 @@ export default class VictoryLine extends React.Component {
   static defaultTransitions = {
     onExit: {
       duration: 500,
-      before: (datum) => ({ y: datum.y })
+      before: (datum) => ({ y: datum.y }),
+      beforeClipPathWidth: (data, child, exitingNodes) => {
```
boygirl: Is this identical to what you are using in VictoryArea? Can you extract this code into a helper. The code for the default could even live in `transition.js` we kept the default `before` / `after` functions here for the others because they were different for each type of component. If `beforeClipPathWidth` etc is always the same you can keep the code with the rest of the transition stuff.

src/components/victory-stack/victory-stack.js
```diff
@@ -413,18 +424,33 @@ export default class VictoryStack extends React.Component {
...
+  renderGroup(children, modifiedProps, style) {
+    const { clipPathComponent } = modifiedProps;
+    const barWidth = style.data.width;
+    const clipComponent = React.cloneElement(clipPathComponent, Object.assign(
+      {},
+      {
+        padding: modifiedProps.padding,
+        clipId: modifiedProps.clipId,
+        barWidth,
+        clipWidth: modifiedProps.clipWidth + barWidth * 2 || modifiedProps.width + barWidth * 2,
```
boygirl: `VictoryStack` shouldn't have to know about barWidth. It also works with area...

src/components/victory-area/helper-methods.js
```diff
@@ -6,6 +6,13 @@ import Scale from "../../helpers/scale";
 
 export default {
 
+  getScale(props, fallbackProps) {
+    const modifiedProps = Helpers.modifyProps(props, fallbackProps);
+    const {scale} = this.getCalculatedValues(modifiedProps);
```
boygirl: I would prefer if getScale() were broken out into its own method and then used in `getCalculatedValues` instead of the other way around so we can avoid unnecessary calculations

src/components/victory-area/victory-area.js
chilijung: oops, forget this one...

src/components/victory-bar/helper-methods.js
```diff
@@ -24,6 +24,22 @@ export default {
     };
   },
 
+  getBarWidth(props) {
+    const {style, width, data} = props;
+    const padding = props.padding.left || props.padding;
+    let barWidth = style && style.width;
+
+    if (!barWidth) {
+      if (data.length === 0) {
```
boygirl: this can be simplified to a ternary return

src/components/victory-bar/victory-bar.js
```diff
@@ -450,21 +463,36 @@ export default class VictoryBar extends React.Component {
     );
   }
 
-  renderGroup(children, style) {
+  renderGroup(children, modifiedProps, style) {
+    const { clipPathComponent } = modifiedProps;
+    const barWidth = BarHelpers.getBarWidth(modifiedProps);
+    const clipComponent = React.cloneElement(clipPathComponent, Object.assign(
+      {},
```
boygirl: Use lodash `assign` instead of `Object.assign` for perf

src/components/victory-group/victory-group.js
```diff
@@ -440,18 +451,33 @@ export default class VictoryGroup extends React.Component {
...
+  renderGroup(children, modifiedProps, style) {
+    const { clipPathComponent } = modifiedProps;
+    const barWidth = style.data.width;
+    const clipComponent = React.cloneElement(clipPathComponent, Object.assign(
+      {},
+      {
+        padding: modifiedProps.padding,
+        clipId: modifiedProps.clipId,
+        barWidth,
+        clipWidth: modifiedProps.clipWidth + barWidth * 2 || modifiedProps.width + barWidth * 2,
```
boygirl: VictoryGroup should not have to know about barWidth

src/components/victory-line/helper-methods.js
```diff
@@ -6,6 +6,13 @@ import Scale from "../../helpers/scale";
 
 export default {
 
+  getScale(props, fallbackProps) {
+    const modifiedProps = Helpers.modifyProps(props, fallbackProps);
+    const {scale} = this.getCalculatedValues(modifiedProps);
```
boygirl: `getCalculatedValues` should rely on `getScale` not the other way around to avoid unnecessary calculations.

src/components/victory-stack/victory-stack.js
chilijung: this is fixed in this commit https://github.com/FormidableLabs/victory-chart/pull/333/commits/99e7af3b863494df1eaf5a38da7cc35390ed3d59

src/components/victory-area/helper-methods.js
chilijung: fixed https://github.com/FormidableLabs/victory-chart/pull/333/commits/87ac4a273fdebcd630261e51fa624c10b65b2ac7 , and https://github.com/FormidableLabs/victory-chart/pull/333/commits/e904b1f498ce98099e44a70b2edf58b6169c066e

src/components/victory-bar/helper-methods.js
chilijung: fixed https://github.com/FormidableLabs/victory-chart/pull/333/commits/7b93994cc504c0ae24a8d65ec8a7b9b02bf91b2d

src/components/victory-bar/victory-bar.js
chilijung: fixed https://github.com/FormidableLabs/victory-chart/pull/333

src/components/victory-group/victory-group.js
chilijung: fixed https://github.com/FormidableLabs/victory-chart/pull/333/commits/99e7af3b863494df1eaf5a38da7cc35390ed3d59

src/components/victory-line/helper-methods.js
chilijung: fixed https://github.com/FormidableLabs/victory-chart/pull/333/commits/cc39480684bc528dab167b5028d3a7de1111d9ca

src/components/victory-line/victory-line.js
chilijung: VictoryLine and VictoryArea is slightly different now.

boygirl: It looks like they are slightly different because your two `getScale` methods return differently. I think the static methods should return the same thing in all the components. I think in VictoryArea this might mean calculating the domain twice, but I think it's worth it for the code cleanliness. And I think it can be refactored so that doesn't happen with a little change to `helpers/scale` I know I'm very picky in this regard, and I am happy to jump in on refactoring that part if you want to move on.

boygirl:
@chilijung I noticed a little detail about clipPath: I think maybe it needs a little bit of padding

screen shot 2016-08-10 at 11 40 06 pm

chilijung:
@boygirl Yeah! How much pixels do you think to remain for paddings? 5 or 10?

boygirl:
I'm also seeing this happen, but only when nested inside chart. It doesn't happen on line alone:

bouncing-clip

boygirl:
@chilijung I think it should be a prop on clipPath. For now it can default to 5px, and we can figure out if we can calculate something smarter from styles etc in a new ticket

chilijung:
OK! I'll add it!

chilijung:
Add clipPadding, and animation problem fixed!

PR #357, opened 2016-08-25
Feature/Fix label overlap issue.

nightwolfdu:
Hello! 
It's partial fix 'label overlapping' , in situation when axis.size < sum(labels.size).
And 'text size' helper, that predict VictoryLabel size based on text and font-style.
I take out filtering code from "renderGridAndTicks" into individual function "fixLabelOverlap", that filter 
"renderGridAndTicks" result.

We discuss about it in Gitter, recently.

> Lauren @boygirl авг. 11 20:05
@deathmandu we could maybe approximate the size of the text based on length, number of lines, font size, padding etc. and filter it in renderGridAndTicks, but it wouldn't be exact. If we went this route, I would want to add a static method to VictoryLabel like getLabelDimensions, and call it in renderGridAndTicks. Option 2 is not viable for us since we use the same code base for victory-native
@traverspinkerton you should style axis stuff on VictoryAxis like...

### DIFF
--- a/src/components/victory-axis/victory-axis.js
+++ b/src/components/victory-axis/victory-axis.js
@@ -1,8 +1,8 @@
-import { assign, defaults, isFunction, partialRight } from "lodash";
 import React, { PropTypes } from "react";
+import { assign, defaults, isFunction, partialRight } from "lodash";
 import {
   PropTypes as CustomPropTypes, Helpers, Events, VictoryTransition, VictoryLabel,
-  VictoryContainer, VictoryTheme, Line
+  VictoryContainer, VictoryTheme, Line, TextSize
 } from "victory-core";
 import AxisHelpers from "./helper-methods";
 import Axis from "../../helpers/axis";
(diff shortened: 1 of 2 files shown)


boygirl:
@nightwolfdu thanks for this improvement! I'm about to head out for a long weekend, but I'll review thoroughly on Monday. One bit of immediate feedback is that the textsize helper should probably go in the victory-core repo. I think it could be broadly useful for anything with a label.

nightwolfdu:
Collapse into 1 commit.

boygirl:
@nightwolfdu I just released your TextSize changes in `victory-core@6.0.0` so you can depend on it for this branch. I'm mostly out of the office today, but I can get this branch in tomorrow morning.

nightwolfdu:
@boygirl  Collapse in one commit, fetch & rebase on master, fix conflict.

boygirl:
@nightwolfdu this looks great! Thanks for writing this enhancement. This will go in the release I have planned for Monday. 

One thing I noticed is that the solution is limited to evenly spaced axis labels. I will open a new issue though, because I think it will be complicated to solve.

labeloverlap

PR #358, opened 2016-08-25
Remove colon from the VictoryScatter @examples annotation.

rawrmonstar:
I noticed the documentation examples for the `<VictoryScatter>` label prop has an extra colon.

This PR removes the colon from the `@examples` notation so that generated docs don't render with an extra colon.

image

cc/ @ebrillhart


ebrillhart:
@rawrmonstar good catch! LGTM!

PR #410, opened 2016-10-31
Pan & Zoom

imranolas:
A naive first pass at implementing pan and zoom. This approach simply wraps the top level chart component with VictoryZoom which handles the events and updates the domain accordingly.

There is a still (a lot) more to be done but it would be great to get some eyes on this in the meantime.

relies on https://github.com/FormidableLabs/victory-core/pull/165
relies on https://github.com/FormidableLabs/victory-core/pull/166

/cc @jevakallio


src/components/victory-zoom/victory-zoom.js
```diff
@@ -0,0 +1,135 @@
...
+            const nextXDomain = scale(1 + (deltaY / 100), xBounds, x);
+            this.setState({domain: {x: nextXDomain}});
+          });
+        }
+      }
+    }];
+
+    const chart = React.Children.only(this.props.children);
+    const nextProps = assign({
+      events,
```
boygirl: We should unshift these events onto whatever events might exist on the chart instance.

src/components/victory-zoom/victory-zoom.js
```diff
@@ -0,0 +1,135 @@
...
+        onMouseUp: () => {
+          this.isPanning = false;
+        },
+        onMouseLeave: () => {
+          this.isPanning = false;
+        },
+        onMouseMove: (evt) => {
+          const clientX = evt.clientX;
+          if (this.isPanning) {
+            requestAnimationFrame(() => {
```
boygirl: Any possibility of reconciling this this with our existing animation model? Could we just force a new domain on the cloned chart, and let the props change trigger animations?

src/components/victory-zoom/victory-zoom.js
```diff
@@ -0,0 +1,135 @@
...
+    children: PropTypes.node,
+    domain: PropTypes.array
+  }
+
+  componentWillMount() {
+    this.getChartRef = (chart) => { this.chartRef = chart; };
+  }
+
+  render() {
+    const events = [{
```
boygirl: We have a concept of `defaultEvents` that get picked up from sub components like `labelComponent` `containerComponent` etc. Maybe this could become `VictoryZoomContainer` which would render the containing `svg` and be responsible for default events. This would get rid of the `refs` issues, and allow zoom to be used for components other than chart. I _think_ the current events system can support this if we explicitly pick up state changes on parent when figuring out domain etc.

boygirl:
@imranolas Related issue
https://github.com/FormidableLabs/victory/issues/386

src/components/victory-zoom/victory-zoom.js
imranolas: We do clone the chart with a new domain on mouseMove. `requestAnimationFrame` in this case throttles the event stream so as not to pummel the render loop.

We actually don't want to animate between domains as the update should appear immediately. Currently if animations are set we do get some odd behaviour, so we need a convenient way to disable animating on domain changes.

src/components/victory-zoom/victory-zoom.js
imranolas: @jevakallio and I did discuss making this a container. The issue we have is that props for the children are calculated in `VictoryChart` or `VictoryGroup`. If we're to move this to a container it's not clear how to pass the new domain up the hierarchy to the parent.

I think semantically it makes sense for zoom to be the root component as it's behaviour should affect the entire chart rather portions of it. It also seems to be a good place to render a brush component at a later stage. Perhaps instead we could define a common interface for any component rendered within `VictoryZoom`, eg. target bounds for the svg could be encapsulated in a method on both `VictoryChart` and `VictoryGroup, so as not to expose the refs directly.

boygirl:
@imranolas this is looking great. The one sort of behavioral oddity I noticed is that if animations are defined, the pinch zoom is not instantaneous. I `npm-link`'d Ken's global animation timer work, and still no dice. I wonder whether something in `VictoryTransition` will need to change.

boygirl:
Also, the clip-path cutoff is inconsistent between the x and y axes 
screen shot 2016-11-07 at 11 09 38 am

boygirl:
@imranolas This is looking awesome! How would you feel about releasing this sooner and addressing the two issues I mention above separately. I would love to have it out and documented by the DF release. I saw your docs PR, but I think this feature also deserves a dedicated guide, and an example on the front page.

src/components/victory-chart/helper-methods.js
```diff
@@ -19,7 +19,7 @@ export default {
     };
 
     if (axisComponents.dependent.length === 0 && axisComponents.independent.length === 0) {
-      return [defaultAxes.independent, defaultAxes.dependent].concat(childComponents);
+      return childComponents.concat([defaultAxes.independent, defaultAxes.dependent]);
```
imranolas: Panning though a data set that crosses the y axis sometimes obscures values on that axis, particularly for area charts. It's arguably more useful to always see the corresponding values. Was there a reason for the axes being rendered under data components?

imranolas:
@boygirl The global timer has now been integrated so the issue with animations will be handled correctly once https://github.com/FormidableLabs/victory-core/pull/166 has landed.

I have couple more updates to make to the docs but if you are happy we can release this for now and the remaining tasks can be added to the backlog.

src/components/victory-chart/helper-methods.js
boygirl: I think it was a preference to have data appear above gridlines by default. I see your point about tick labels though. That does seem more important than whatever is going on with the gridlines.

PR #462, opened 2017-04-19
Properly position tooltip in voronoi containers for orientations other than “top”

GreenGremlin:
For orientations other than "top" tooltips inside a voronoi container currently spill outside the chart container. This fix ensures they are properly contained.


GreenGremlin:
Is there anything I can do to help get this merged?

PR #469, opened 2017-05-11
VictoryCursorContainer

chrisbolin:
Implements VictoryCursorContainer (https://github.com/FormidableLabs/victory/issues/514)

demo at http://localhost:3000/#/cursor-container

Open issues:
- Where should the label go for 1-dimensional cursors? (with mouse [current implementation], outside of chart, middle of chart, something else?)
- createContainer("cursor", "voronoi") works pretty well, but is awkward when `labels` is used (both containers try to create labels, but they are in different places and for different values)

### DIFF
--- a/src/components/containers/victory-cursor-container.js
+++ b/src/components/containers/victory-cursor-container.js
@@ -9,6 +9,15 @@ export const cursorContainerMixin = (base) => class VictoryCursorContainer exten
   static displayName = "VictoryCursorContainer";
   static propTypes = {
     ...VictoryContainer.propTypes,
+    cursorLabel: PropTypes.func,
+    cursorLabelComponent: PropTypes.element,
+    cursorLabelOffset: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.shape({
+        x: PropTypes.number,
+        y: PropTypes.number
+      })
+    ]),
     defaultCursorValue: PropTypes.oneOfType([
       PropTypes.number,
       PropTypes.shape({
@@ -17,22 +26,13 @@ export const cursorContainerMixin = (base) => class VictoryCursorContainer exten
       })
     ]),
     dimension: PropTypes.oneOf(["x", "y"]),
-    labelComponent: PropTypes.element,
-    labelOffset: PropTypes.oneOfType([
-      PropTypes.number,
-      PropTypes.shape({
-        x: PropTypes.number,
-        y: PropTypes.number
-      })
-    ]),
-    labels: PropTypes.func,
     onChange: PropTypes.func,
     standalone: PropTypes.bool
   };
   static defaultProps = {
     ...VictoryContainer.defaultProps,
-    labelComponent: <VictoryLabel/>,
-    labelOffset: {
+    cursorLabelComponent: <VictoryLabel/>,
+    cursorLabelOffset: {
       x: 5,
       y: -10
     },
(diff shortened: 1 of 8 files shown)


chrisbolin:
### VictoryCursorContainer

may-11-2017 19-59-38

chrisbolin:
### `createContainer("cursor", "voronoi")`

may-11-2017 19-59-06

src/components/containers/victory-cursor-container.js
```diff
@@ -0,0 +1,123 @@
...
+import React from "react";
+import { VictoryContainer, VictoryLabel, Line, Selection, Helpers } from "victory-core";
+import { isNumber } from "lodash";
+import CursorHelpers from "./cursor-helpers";
+
+export const cursorContainerMixin = (base) => class VictoryCursorContainer extends base {
+  static displayName = "VictoryCursorContainer";
+  static propTypes = {
+    ...VictoryContainer.propTypes,
+    defaultCursorValue: PropTypes.oneOfType([PropTypes.number, PropTypes.object]),
```
boygirl: Prefer `PropTypes.shape(x... y...)` to `PropTypes.object`

src/components/containers/victory-cursor-container.js
```diff
@@ -0,0 +1,123 @@
...
+        x1: domainCoordinates.x[0],
+        x2: domainCoordinates.x[1],
+        y1: cursorCoordinates.y,
+        y2: cursorCoordinates.y
+      }));
+    }
+    return newElements;
+  }
+
+  getTooltip(props) {
```
boygirl: `getLabel`

src/components/containers/victory-cursor-container.js
```diff
@@ -0,0 +1,123 @@
...
+      x: scale.x(cursorPosition.x),
+      y: scale.y(cursorPosition.y)
+    };
+    if (labels) {
+      newElements.push(React.cloneElement(labelComponent, {
+        x: cursorCoordinates.x + 5,
+        y: cursorCoordinates.y - 10,
+        text: Helpers.evaluateProp(labels, cursorPosition, true),
+        active: true,
+        key: "cursor-tooltip"
```
boygirl: "cursor-label"

src/components/containers/victory-cursor-container.js
chrisbolin: yup!

src/components/containers/victory-cursor-container.js
```diff
@@ -0,0 +1,123 @@
...
+    const newElements = [];
+    const domainCoordinates = Selection.getDomainCoordinates(scale, domain);
+
+    const cursorCoordinates = {
+      x: scale.x(cursorPosition.x),
+      y: scale.y(cursorPosition.y)
+    };
+    if (labels) {
+      newElements.push(React.cloneElement(labelComponent, {
+        x: cursorCoordinates.x + 5,
```
boygirl: hard coded positioning here seems wrong. Maybe a `labelOffset` prop that can take a single value, or values for x and y.

boygirl: The values you have here can be the defaults

src/components/containers/victory-cursor-container.js
```diff
@@ -0,0 +1,123 @@
...
+      scale, domain, dimension, labelComponent, labels, cursorComponent
+    } = props;
+    const cursorPosition = this.getCursorPosition(props);
+
+    if (!cursorPosition) { return []; }
+
+    const newElements = [];
+    const domainCoordinates = Selection.getDomainCoordinates(scale, domain);
+
+    const cursorCoordinates = {
```
boygirl: You're already getting the svg coordinates of the mouse position and transforming those into data coordinates in your helpers. You could just add the svg coordinates to the mutation in `onMouseMove`

boygirl:
Oh! actually, I just noticed... do `labels` and `labelComponent` conflict when you use `createContainer` with "voronoi" and "cursor"? How would you define independent values?

src/components/containers/victory-cursor-container.js
chrisbolin: thank you! i felt the same way

chrisbolin:
@boygirl YES, you are correct. When `createContainer` is used, they will both reference the same `labels` prop. That's why the example above (top-left under `createContainer("cursor", "voronoi")`) has two tooltips.

We could rename the props. This is definitely the simplest solution, but also is kind of skirting the problem. But hey, being specific (like with `voronoiLabels` and `cursorLabelComponent`) does make things more clear.

I'm not sure exactly what another solution would look like. `createContainer` literally creates a single class that implements the hybrid container.

boygirl:
In that case, let's go with `cursorLabels` and `cursorLabelComponent`. We can leave `VictoryVoronoiContainer` as is to avoid a breaking change

src/components/containers/victory-cursor-container.js
chrisbolin: totally understand. i think i'll keep it this way though, because it handles getting the position when `defaultCursorValue` (which can be both number and object) is used (i.e. there is no `onMouseMove`).

boygirl: 👍

chrisbolin:
@boygirl addressed those issues. waiting for CI to pass before I do anything.

also, i went with `cursorLabel` (singular) instead of `cursorLabels`, as a cursor can only have one label, and cursorLabelComponent was singular. Totally happy to do plural `cursorLabels` if you think that would be more consistent with other components.

boygirl:
@chrisbolin looks good to me! Approved!

boygirl:
I'm about to do a quick release, so I'm going to go ahead and merge this :)

PR #581, opened 2018-04-13
add react-fast-compare

chrisbolin:
- add `react-fast-compare`. use in place of `lodash.isEqual`
- add demo that uses `<VictoryLabel/>` and triggers a `victory-core` `shouldComponentUpdate`, which would have caught https://github.com/FormidableLabs/victory/issues/964


chrisbolin:
caught by `comma-dangle`, will fix

